PR #5, opened 2019-12-06
add solution hw53. i'd redone it 3 times, before read that readString…

karsv:
… return String...

### DIFF
--- a/src/main/java/core/basesyntax/FileWork.java
+++ b/src/main/java/core/basesyntax/FileWork.java
@@ -19,7 +19,6 @@ public class FileWork {
 
     public String[] readFromFile(String fileName) throws IOException {
         String stringFile = Files.readString(Paths.get(fileName));
-        ;
         String[] temp = stringFile.toLowerCase().replaceAll("[^a-z]+", " ").split(" ");
         List<String> list = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


src/test/java/core/basesyntax/FileWorkTest.java
```diff
@@ -16,15 +18,15 @@
             new String[]{"wall", "wave", "width", "world", "www"};
 
     @Test
-    public void readFromEmptyFile() {
+    public void readFromEmptyFile() throws IOException {
```
Tarabarnik: you are not allowed to change tasks. please, create solution, which doesn't throw an Exception

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,20 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+    final String firstLetter = "w";
```
Tarabarnik: move this constant to class please

karsv: i made it private static final? is it wright? B'cos, how i understand, it was at class

PR #7, opened 2019-12-06
hw 5 3

SergiyAgeev:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,31 @@
...
+
+        try {
+            File file = new File(fileName);
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(file));
+            String temp;
+
+            while ((temp = bufferedReader.readLine()) != null) {
+                String[] result = temp.toLowerCase().replaceAll("[^a-z]", " ").split(" ");
+                for (String s : result) {
+                    if (s.startsWith("w")) {
```
Tarabarnik: make "w" as a constant of a class

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,31 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    final String mostImportantChar = "w";
```
Tarabarnik:
```suggestion
    private static final String KEYWORD = "w";
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,31 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    private static final String CONSTANT = "w";
```
Tarabarnik: naming should be more informative. what if we had 10 constants in the class?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,31 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> resultList = new ArrayList<>();
+
+        try {
+            File file = new File(fileName);
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(file));
```
Tarabarnik: readers should be closed after we don't need them. there is a method called `close()`. but to close them automatically, we should use try with resources

src/main/java/core/basesyntax/FileWork.java
Tarabarnik: what about this constant name??

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,48 @@
...
+                        if (s.startsWith(CONSTANT)) {
+                            resultList.add(s);
+                        }
+                    }
+                }
+            }
+
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        } catch (IOException e) {
```
Tarabarnik: when catch blocks are same, you may collapse them

PR #9, opened 2019-12-07
added hw 5.3 solution

programmer28:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,34 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String resultLine = "";
+        try {
```
Tarabarnik: can we use try with resources?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,34 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String resultLine = "";
```
Tarabarnik: try to use StringBuilder

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,34 @@
...
+        } catch (FileNotFoundException e) {
+            System.out.println(e);
+        } catch (IOException e) {
+            System.out.println(e);
+        }
+        resultLine = resultLine.replaceAll("[^(\\sa-zA-Z)]", "");
+        String[] words = resultLine.split(" ");
+        List<String> list = new ArrayList<>();
+        for (String word : words) {
+            if (word.toLowerCase().startsWith("w")) {
```

programmer28: I`ve changed it

PR #11, opened 2019-12-07
I guess I will make a lot of changes ^_^

MrNopi:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,30 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> arr = new ArrayList<>();
+
+        File file = new File(fileName);
```
Tarabarnik: try to create a solution without this variable

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,30 @@
...
+        if (file.length() == 0) {
+            return new String[]{};
+        }
+        Path path = Paths.get(fileName);
```
Tarabarnik: same

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,30 @@
...
+        try {
+            List<String> array = Files.readAllLines(path);
+            for (String value : array) {
+                String[] tmp = value.toLowerCase().split(" ");
+                for (String s : tmp) {
+                    if (s.startsWith("w")) {
```

MrNopi: > 
> 
> make "w" constant of a class

I will also make regex constant, is it fine?

Tarabarnik: as you wish :) it's optional

src/main/java/core/basesyntax/FileWork.java
MrNopi: Which one? Without File or without list?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
```
Tarabarnik: we shouldn't use multiple imports like `import java.util.*;`

src/main/java/core/basesyntax/FileWork.java
Tarabarnik: what about this comment?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,29 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+
+    private static final String W = "w";
```
Tarabarnik: will this name of constant be informative, if you change its value later?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,27 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+
+    private static final String START_OF_WORDS = "w";
+    private static final String REGEX_NON_WORDS = "[^a-z+]";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> arr = new ArrayList<>();
```
Tarabarnik: variable names should be full and informative. 
https://mate-academy.github.io/style-guides/java/java.html#s5.2.5-non-constant-field-names

PR #12, opened 2019-12-07
Added solution for hw 5.3

ElvenNurse:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,27 @@
...
+        String[] result = new String[0];
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            StringBuilder searchedWords = new StringBuilder();
+            while ((line = bufferedReader.readLine()) != null) {
+                String[] buffuredWords = line.toLowerCase()
+                        .replaceAll("[^a-z]", " ")
+                        .trim().split(" ");
+                for (String word: buffuredWords) {
+                    if (word.startsWith("w")) {
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,28 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String[] result = new String[0];
+        final String firstChar = "w";
```
Tarabarnik: move this constant as a class field, please

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,28 @@
...
+                String[] bufferedWords = line.toLowerCase()
+                        .replaceAll("[^a-z]", " ")
+                        .split(" ");
+                for (String word: bufferedWords) {
+                    if (word.startsWith(firstChar)) {
+                        searchedWords.append(word).append(" ");
+                    }
+                }
+            }
+            if (!searchedWords.toString().isEmpty()) {
```
Tarabarnik: try to solve without this check

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,28 @@
...
+                result = searchedWords.toString().split(" ");
+                Arrays.sort(result);
+            }
+        } catch (IOException e) {
+            System.out.println("Sorry =(");
```
Tarabarnik: not much informative :)

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,31 @@
...
+                String[] bufferedWords = line.toLowerCase()
+                        .replaceAll("[^a-z]", " ")
+                        .split(" ");
+                for (String word: bufferedWords) {
+                    if (word.startsWith(FIRST_CHAR)) {
+                        searchedWords.add(word);
+                    }
+                }
+            }
+            result = new String[searchedWords.size()];
```
Tarabarnik: ArrayList has a method called `toArray()`, so you don't need to create this block

Tarabarnik: https://www.geeksforgeeks.org/arraylist-array-conversion-java-toarray-methods/

ElvenNurse: toArray(), in this case, works only if we cast ArrayList to String with generic... I'm trying to avoid using generic as much as I can, cause it's our next lesson... but OK, I surrender :-p

PR #21, opened 2019-12-09
added hw-5-3-solution

AlinaPodoliaka:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,36 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String str = "";
```
Tarabarnik: try to use StringBuilder

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,36 @@
...
+        try {
```
Tarabarnik: try with resources?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,36 @@
...
+        }
+        str = str.replaceAll("[^\\s a-z A-Z]", "");
+        String[] words = str.split(" ");
+        List<String> list = new ArrayList<>();
+        for (int i = 0; i < words.length; i++) {
+            if (words[i].toLowerCase().startsWith("w")) {
+                list.add(words[i].toLowerCase());
+            }
+        }
+        if (list.isEmpty()) {
```
Tarabarnik: can we crate a solution without this check?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,36 @@
...
+        } catch (FileNotFoundException e) {
+            System.out.println(e);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        str = str.replaceAll("[^\\s a-z A-Z]", "");
+        String[] words = str.split(" ");
+        List<String> list = new ArrayList<>();
+        for (int i = 0; i < words.length; i++) {
+            if (words[i].toLowerCase().startsWith("w")) {
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,36 @@
...
+                list.add(words[i].toLowerCase());
+            }
+        }
+        if (list.isEmpty()) {
+            return new String[0];
+        }
+        Collections.sort(list);
+        String[] result = new String[list.size()];
+        for (int i = 0; i < result.length; i++) {
```
Tarabarnik: try to use toArray() method of List instead of custom loops

src/main/java/core/basesyntax/FileWork.java
AlinaPodoliaka: yes,we can. Sorry, it was late night, when I finished hw, so I add this check. But now I see, that we can figure without it.

src/main/java/core/basesyntax/FileWork.java
AlinaPodoliaka: done

src/main/java/core/basesyntax/FileWork.java
AlinaPodoliaka: done

PR #23, opened 2019-12-09
Added Mate HW5.3 Solution

AlexanderSashchenko:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,31 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        String text = null;
+        try {
+            text = Files.readString(Paths.get(fileName));
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        if (text.length() == 0) {
```
Tarabarnik: try to create a solution without this check

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,31 @@
...
+            return new String[0];
+        }
+        String[] words = text.toLowerCase().replaceAll("[^a-z]+", " ").split(" ");
+        StringBuilder newContainer = new StringBuilder();
+        for (String word : words) {
+            if (word.charAt(0) == LETTER) {
+                newContainer.append(word + " ");
+            }
+        }
+        if (newContainer.length() == 0) {
```
Tarabarnik: try to create a solution without this check, too

src/main/java/core/basesyntax/FileWork.java
AlexanderSashchenko: Fixed.

src/main/java/core/basesyntax/FileWork.java
AlexanderSashchenko: Fixed.

PR #24, opened 2019-12-12
added readFromFile solution

TarabanDmitry:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,27 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+
+        List<String> line = new ArrayList<String>();
+        try {
+            File file = new File(fileName);
```
Tarabarnik: try to solve without creating File

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,27 @@
...
+            FileReader fileReader = new FileReader(file);
+            BufferedReader bufferedReader = new BufferedReader(fileReader);
```
Tarabarnik: readers should be closed, so try to use `try with resources`

PR #26, opened 2020-04-02
Hw 5.3

fruts:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
 /**
  * <p>Дано файл, потрібно прочитати його вміст і вибрати всі слова що починаються на `w`.
  * Результат повернути у вигляді відсортованого масиву (за замовчуванням). Всі слова повинні
  * бути в нижньому регістрі. У випадку якщо таких слів не знайдено повернути пустий масив.
  * Приклад: Width world Wide web
  * Результат: web wide width world</p>
  */
+
 public class FileWork {
+    public static final String BIG_W = "W";
```
Tarabarnik, line 17:
```suggestion
    public static final String CHARACTER = "W";
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
+    public static final String SMALL_W = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        File file = new File(fileName);
+        StringBuilder scannedWords = new StringBuilder();
+        try (Scanner scan = new Scanner(file)) {
+            while (scan.hasNext()) {
+                String oneString = scan.next();
+                if (oneString.startsWith(SMALL_W) || oneString.startsWith(BIG_W)) {
```
Tarabarnik, line 26: try to solve with a single `startsWith` call

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
+                    scannedWords.append(oneString
+                            .toLowerCase()
+                            .replaceAll("[^a-z]", ""))
```
Tarabarnik, line 29: do you really need this `replaceAll`?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
+                            .append(" ");
+                }
+            }
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
Tarabarnik, line 34: don't ignore exceptions. see the checklist

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,38 @@
...
+                    scannedWords.append(oneString.replaceAll("\\W", "")).append(" ");
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Неверно указано имя файла, или его не существует", e);
+        }
+        return scannedWords.length() > 0
+                ? Arrays.stream(scannedWords.toString().trim()
+                .split(" ")).sorted().toArray(String[]::new)
+                : new String[0];
```
Tarabarnik, line 35:
```suggestion
        return scannedWords.toString().split(" ");
```
this should work, if not - change your code. sorry, didn't notice at first review

fruts: your option does not work

Tarabarnik: yes, but they won't be sorted. try to use Arrays.sort() with a result of that split method

PR #27, opened 2020-04-02
Homework 4_3

ZubenkoHalyna:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,39 @@
...
+    private static final String CHAR_AND_SPACE_ONLY_REGEXP = "[^A-z ]";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder words = new StringBuilder();
+        try (BufferedReader br = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = readStringFromFile(br)) != null) {
+                for (String word : line.toLowerCase().replaceAll(CHAR_AND_SPACE_ONLY_REGEXP, "")
+                        .split(" ")) {
```
dmytro-hryhoruk, line 25: let's declare String[] for it

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,39 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        if (words.length() != 0) {
+            String[] array = words.toString().split(" ");
+            Arrays.sort(array);
+            return array;
+        } else {
+            return new String[0];
+        }
```
dmytro-hryhoruk, line 40: do we need this check?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,39 @@
...
     }
+
+    private String readStringFromFile(BufferedReader br) {
+        try {
+            return br.readLine();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
```
dmytro-hryhoruk, line 50: do we need such a small method?

src/main/java/core/basesyntax/FileWork.java
ZubenkoHalyna: Yes, otherwise StringBuilder.toString().split(" ") return {""} and tests fail.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,27 @@
...
+                        .split(SPLIT_TEXT)) {
+                    if (word.startsWith(START_OF_WORD)) {
+                        words.append(word).append(SPLIT_TEXT);
+                    }
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        String[] array = words.toString().split(SPLIT_TEXT);
```
dmytro-hryhoruk, line 35: we don't need constant for whitespace

PR #31, opened 2020-04-02
Added solution.

Merkald:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,46 @@
...
-        return null;
+        StringBuilder stringBuilder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName));) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                stringBuilder.append(line.toLowerCase().replaceAll("\\p{Punct}", ""));
+                stringBuilder.append(" ");
+            }
+        } catch (IOException e) {
+            return new String[0];
```
dmytro-hryhoruk, line 29: you should re-throw exception, don't ignore them

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,46 @@
...
+            if (stringBuilder.charAt(0) == WORD_START_AT) {
+                result.append(stringBuilder.subSequence(0, stringBuilder.indexOf(" "))).append(" ");
+            }
+            stringBuilder.delete(0, stringBuilder.indexOf(" ") + 1);
+        }
+        if (result.length() != 0) {
+            String[] finalResult = result.toString().split(" ");
+            Arrays.sort(finalResult);
+            return finalResult;
+        } else {
```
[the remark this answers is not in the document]
Merkald: the condition is not fulfilled without this check returns String[1]="" And tests are fails.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,47 @@
...
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder stringBuilder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName));) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                stringBuilder.append(line.toLowerCase().replaceAll(ALL_PUNCTUATION, ""));
+                stringBuilder.append(" ");
+            }
+        } catch (RuntimeException | IOException e) {
```
dmytro-hryhoruk, line 29: do we need to catch any RuntimeException here?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,47 @@
...
+            throw new RuntimeException();
```
dmytro-hryhoruk, line 30: add some informative message to exception

src/main/java/core/basesyntax/FileWork.java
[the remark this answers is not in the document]
dmytro-hryhoruk: Yes, try to use List<String> instead of StringBuilder here

PR #32, opened 2020-04-02
added solution

AlexeyBavika:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,30 @@
...
     public String[] readFromFile(String fileName) {
-        return null;
+        File file = new File(fileName);
+        StringBuilder wordsStartWithW = new StringBuilder();
+        try (Scanner scanner = new Scanner(file)) {
+            if (!scanner.hasNext()) {
+                return new String[0];
+            }
+            while (scanner.hasNext()) {
+                String word = scanner.next();
```
```suggestion
                String word = scanner.next().toLowerCase();
```

dmytro-hryhoruk: and then you don't have to check for Bigger W

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,30 @@
...
+                if (word.startsWith(BIG_W) || word.startsWith(SMALL_W)) {
+                    wordsStartWithW.append(word.toLowerCase().replaceAll("[^a-zA-z]", "") + " ");
+                }
+            }
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        }
+        String[] sortedWords = wordsStartWithW.toString().trim().split(" ");
+        if (sortedWords.length <= 1) {
+            return new String[0];
```
dmytro-hryhoruk, line 37: do we need this check?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,26 @@
...
+                return new String[0];
+            }
+            while (scanner.hasNext()) {
+                String word = scanner.next().toLowerCase();
+                if (word.startsWith(SMALL_W)) {
+                    wordsStartWithW.append(word.replaceAll("[^a-zA-Z]", "") + " ");
+                }
+            }
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
dmytro-hryhoruk, line 32: you should re-throw exception, not ignore them

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,26 @@
...
+        }
+        String[] sortedWords = wordsStartWithW.toString().trim().split(" ");
+        Arrays.sort(sortedWords);
+        return sortedWords.length > 1 ? sortedWords : new String[0];
```
dmytro-hryhoruk, line 36: and still, do we need this check?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,26 @@
...
+    private static final String SMALL_W = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        File file = new File(fileName);
+        StringBuilder wordsStartWithW = new StringBuilder();
+        try (Scanner scanner = new Scanner(file)) {
+            if (!scanner.hasNext()) {
+                return new String[0];
+            }
```
dmytro-hryhoruk, line 24: do we need this check?

AlexeyBavika:
unfortunately its impossible to remove this check because at the end method split adds space and length = 1 then.

PR #34, opened 2020-04-02
added solution

VitaliiLeshchenko:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,24 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+    public String[] readFromFile(String fileName) throws IOException {
```
dmytro-hryhoruk, line 18: let's handle this exception with try catch

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,24 @@
...
+
+        List<String> lines = Files.readAllLines(Paths.get(fileName));
+        List<String> listResult = new ArrayList<>();
+        for (String line : lines) {
+            String[] splitLine = line.split(" ");
+            for (String word : splitLine) {
+                if (word.charAt(0) == 'w' || word.charAt(0) == 'W') {
```
dmytro-hryhoruk, line 25: do we need to check both of this cases?

src/main/java/core/basesyntax/FileWork.java
VitaliiLeshchenko: 1.added try - catch block.

src/main/java/core/basesyntax/FileWork.java
VitaliiLeshchenko: 2. simplified if condition.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,29 @@
...
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+
+        List<String> lines;
+        try {
+            lines = Files.readAllLines(Paths.get(fileName));
+        } catch (IOException e) {
+            e.printStackTrace();
+            return null;
```
dmytro-hryhoruk, line 25: let's imagine i'm using this method and it's couldn't find a file and throw an exception, but program will continue to work and i will just get the null(not knowing that exception been thrown)

VitaliiLeshchenko: done.
changed "catch" block

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,25 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+
```
dmytro-hryhoruk, line 19: remove empty line

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,25 @@
...
+            String[] splitLine = line.split(" ");
+            for (String word : splitLine) {
+                if (word.toLowerCase().charAt(0) == 'w') {
+                    listResult.add(word.replaceAll("\\W+", "").toLowerCase());
+                }
+            }
+        }
+        String[] result = listResult.toArray(String[]::new);
+        Arrays.sort(result);
+
```
dmytro-hryhoruk, line 37: same

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,25 @@
...
+        List<String> listResult = new ArrayList<>();
+        for (String line : lines) {
+            String[] splitLine = line.split(" ");
+            for (String word : splitLine) {
+                if (word.toLowerCase().charAt(0) == 'w') {
+                    listResult.add(word.replaceAll("\\W+", "").toLowerCase());
+                }
+            }
+        }
+        String[] result = listResult.toArray(String[]::new);
```
dmytro-hryhoruk, line 35: you can use Collections.sort()

PR #35, opened 2020-04-02
Hw 1

6ALLIKA:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,43 @@
...
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder wordsFromFile = new StringBuilder();
+        Path filePath = Paths.get(fileName);
+        List<String> lines = null;
+        try {
+            lines = Files.readAllLines(filePath);
+        } catch (IOException e) {
+            e.printStackTrace();
```
Tarabarnik, line 26: don't ignore excepitons. read the checklist

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,43 @@
...
+        }
+
+        for (String text : lines) {
+            for (String soloText : text.split(" ")) {
+                if (soloText.startsWith("w") || soloText.startsWith("W")) {
```
Tarabarnik, line 31: try to solve with a single check and make 'w' constant

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,43 @@
...
+                    wordsFromFile.append(soloText.toLowerCase()
+                            .replaceAll("[^a-z]", "")).append(" ");
```
Tarabarnik, line 33: do you really need this `replaceAll`?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,43 @@
...
+                }
+            }
+        }
+
+        String[] result = wordsFromFile.toString().split(" ");
+        Arrays.sort(result);
+
+        return wordsFromFile.length() == 0 ? new String[0] : result;
```
Tarabarnik, line 41:
```suggestion
        return result;
```
it should work, I suppose

src/main/java/core/basesyntax/FileWork.java
6ALLIKA: yes? cause we have symbols like . ? etc

src/main/java/core/basesyntax/FileWork.java
6ALLIKA: it would be worked if change some logic

PR #36, opened 2020-04-02
Added solution for HW_5.3

polinakulyk:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +18,16 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        try (Stream<String> lines = Files.lines(Paths.get(fileName))) {
+            List<String> words = lines
+                    .flatMap(line -> Arrays.stream(line.split("\\s+")))
+                    .filter(word -> word.startsWith("W") || word.startsWith("w"))
```
dmytro-hryhoruk, line 24: probably it's better to use toLowerCase() before that and you will have only one check?

dmytro-hryhoruk: create a constant for 'w'

PR #43, opened 2020-04-02
implementation class FileWork

KrohIgor:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,34 @@
...
+                if (word.charAt(0) == charW) {
+                    wordsStartW.append(word.replaceAll("\\W", "")).append(' ');
+                }
+            }
+        }
+        if (wordsStartW.length() > 0) {
+            String[] result = wordsStartW.toString().split(" ");
+            Arrays.sort(result);
+            return result;
+        } else {
```
dmytro-hryhoruk, line 43: do we need else after return ?

KrohIgor: we can without else.

PR #46, opened 2020-04-02
added the method readFromFile()

Andrii-Kutil:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,24 @@
...
-        return null;
+    private static final String FIRST_CHAR = "w";
+
+    public static String[] readFromFile(String fileName) {
+        StringBuilder sb = new StringBuilder();
+        String text = "";
+        try {
+            text = Files.readString(Paths.get(fileName));
+        } catch (IOException e) {
+            e.printStackTrace();
```
dmytro-hryhoruk, line 24: don't ignore exception, re-throw it

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,24 @@
...
  */
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+    private static final String FIRST_CHAR = "w";
+
+    public static String[] readFromFile(String fileName) {
+        StringBuilder sb = new StringBuilder();
+        String text = "";
+        try {
```
dmytro-hryhoruk, line 21: use try-with-resources here, explain why in comment below

dmytro-hryhoruk: ?

PR #48, opened 2020-04-02
internal commit (added solution)

vztot:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,28 @@
...
+                for (String word : splitData) {
+                    word = word.replaceAll(REGEX, "");
+                    if (word.startsWith(LETTER)) {
+                        words.add(word);
+                    }
+                }
+                line = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
dmytro-hryhoruk, line 37: don't ignore the exception, re-throw it

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,28 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> words = new ArrayList<>();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                String[] splitData = line.toLowerCase().split(" ");
+                for (String word : splitData) {
+                    word = word.replaceAll(REGEX, "");
```
dmytro-hryhoruk, line 29: what exactly this regex for? do we need constant for it?

PR #50, opened 2020-04-02
added first solution

tanchak123:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,41 @@
...
+
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+
+    public static String[] readFromFile(String fileName) {
+        final StringBuilder sb = new StringBuilder();
+        List<String> reader;
+        Path filePath = Paths.get(fileName);
+        String [] split = new String[0];
```
Tarabarnik, line 16:
```suggestion
        String[] split = new String[0];
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,41 @@
...
+        try {
+            reader = Files.readAllLines(filePath);
+        } catch (IOException e) {
+            throw new RuntimeException("Причина в том, что " + e);
+        }
+        if (reader.size() > 0) {
+            for (String a : reader) {
+                sb.append(a.toLowerCase() + " ");
```
Tarabarnik, line 24: lets append only words that start from 'w' . BTW, "w" should be a constant, read the checklist

PR #52, opened 2020-04-02
Added solution

aborozdykh:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,23 @@
...
+        try {
+            String string = Files.readString(Paths.get(fileName));
+            String[] stringArray = string.toLowerCase().split(" ");
+            for (String s : stringArray) {
+                if (s.startsWith(PATTERN)) {
+                    sb.append(s.replaceAll("\\W", "")).append(" ");
+                }
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
dmytro-hryhoruk, line 29: do not ignore exception, re-throw it

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,23 @@
...
+        }
+        return sb.length() > 0
```
dmytro-hryhoruk, line 31: do we need this check?

PR #54, opened 2020-04-02
solution

MaksymMatuliak:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,27 @@
...
+                if (i.startsWith(START_OF_WORD)) {
+                    wordsFromW.append(i.replaceAll("\\W", "")).append(" ");
+                }
+            }
+            if (wordsFromW.length() > 0) {
+                result = wordsFromW.toString().split(" ");
+            }
+            Arrays.sort(result);
+        } catch (IOException e) {
+            System.out.println("No files");
```
Tarabarnik, line 35: here you basically do nothing with this exception

MaksymMatuliak: done

PR #55, opened 2020-04-02
Add solution hw-5.3 read-from-file

nevskyy:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,30 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+            return null;
+        }
+        List<String> resultList = new ArrayList<>();
+        for (String sentence : sentences) {
+            String[] splittedSentence = sentence.toLowerCase().split(" ");
+            for (String word : splittedSentence) {
+                if (word.startsWith(FIRST_LETTER)) {
+                    resultList.add(word);
```
Tarabarnik, line 33: can we call `replaceAll` here?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,30 @@
...
+                }
+            }
+        }
+        String[] solution = new String[resultList.size()];
```
Tarabarnik, line 37: is there any method of List, that converts it to Array?

nevskyy: Done as advised.

PR #58, opened 2020-04-03
done

AndrewLinkii:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,24 @@
...
+        StringBuilder words = new StringBuilder();
+        try (Scanner scan = new Scanner(new File(fileName))) {
+            while (scan.hasNext()) {
+                String word = scan.next().toLowerCase();
+                if (word.startsWith("w")) {
+                    words.append(word.replaceAll("[^a-z]", "")).append(" ");
+                }
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println("file not found");
```
Tarabarnik, line 26: throw runtimeexception or return empty array. print of the text is not the exception handling

PR #59, opened 2020-04-03
5.3 var 1

ana-dav:
read from empty file test failed


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,37 @@
...
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+
+    public static final String W = "w";
```
dmytro-hryhoruk, line 10: let's give more informative name

PR #61, opened 2020-04-03
hw 5.3

Djeckkenni:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,29 @@
...
+
+        try {
+            list = Files.readAllLines(filePath);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
+        }
+
+        for (String line : list) {
+            for (String word : line.split(" ")) {
+                if (word.startsWith("w") || word.startsWith("W")) {
```
Tarabarnik, line 31: try to solve with only one check

Djeckkenni: Done

PR #70, opened 2020-04-03
Hw 5.3

Kaijou88:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,49 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+
+        File file = new File(fileName);
+        if (file.length() == 0) {
+            return new String[0];
+        }
+
+        String line = "";
```
Tarabarnik, line 25: Use StringBuilder to avoid concatenation

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,49 @@
...
+
+        for (int i = 0; i < finishArray.length - 1; i++) {
+            for (int j = i + 1; j < finishArray.length; j++) {
+                if (finishArray[i].compareTo(finishArray[j]) > 0) {
+                    String temp = finishArray[i];
+                    finishArray[i] = finishArray[j];
+                    finishArray[j] = temp;
+                }
+            }
+        }
```
Tarabarnik, line 58: try to use Arrays.sort()

Kaijou88: used it

src/main/java/core/basesyntax/FileWork.java
Kaijou88: changed

PR #73, opened 2020-04-03
Done

vovkandrew:
PS I could make some unintentional corrections at the Test file, sorry, tried to fine out why it asked me to throw exceptions

### DIFF
--- a/src/main/java/core/basesyntax/FileWork.java
+++ b/src/main/java/core/basesyntax/FileWork.java
@@ -35,8 +35,7 @@ public class FileWork {
             Collections.sort(listOfWords);
             return listOfWords.toArray(new String[0]);
         } catch (IOException e) {
-            e.printStackTrace();
+            throw new RuntimeException(e);
         }
-        return null;
     }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/core/basesyntax/FileWorkTest.java
```diff
@@ -3,6 +3,8 @@
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.io.IOException;
```
dmytro-hryhoruk, line 6: don't make any changes in tests

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +17,26 @@
...
+            ArrayList<String> listOfWords = new ArrayList<>();
+            for (String i : split) {
+                if (i.startsWith("w")) {
+                    listOfWords.add(i);
+                }
+            }
+            Collections.sort(listOfWords);
+            return listOfWords.toArray(new String[0]);
+        } catch (IOException e) {
+            e.printStackTrace();
```
dmytro-hryhoruk, line 38: we should re-throw exception

PR #77, opened 2020-07-31
ReadFromFile method

andriylagun:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +12,25 @@
...
+    private static final String FIRST_CHAR = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        String fileText = "";
+        try (FileReader fileReader = new FileReader(fileName)) {
+            int oneLetter;
+            while ((oneLetter = fileReader.read()) != -1) {
+                fileText += (char) oneLetter;
+            }
```
dmytro-hryhoruk, line 24: let's use BufferedReader so you will read the whole line per iteration instead of single character

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,24 @@
...
-        return null;
+
+    private static final String FIRST_CHAR = "w";
+
+    public static String[] readFromFile(String fileName) {
+        String fileText = "";
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String oneLine;
+            while ((oneLine = bufferedReader.readLine()) != null) {
+                fileText += oneLine;
```
dmytro-hryhoruk, line 24: let's use StringBuilder for this

PR #79, opened 2020-07-31
Task 3 is done

andromash:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,34 @@
...
+                line = line.replaceAll("[^a-z]", " ");
+                String[] separateWords = line.split("\\s+");
+                for (String word : separateWords) {
+                    if (word.startsWith(KEY)) {
+                        result.add(word);
+                    }
+                }
+            }
+        } catch (IOException e) {
+            System.out.println("No such a file");
```
Sofasmile, line 28: Don’t ignore exceptions. Re-throw them as RuntimeException(e.g. throw new RuntimeException(e)) if you don’t expect your code to fail here, or make necessary actions at the catch block to fix the problem.

PR #81, opened 2020-07-31
hw-5.3

julia-shvabovich:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,26 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder fileContent = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            while (reader.ready()) {
+                fileContent.append(reader.readLine());
+            }
+        } catch (Exception e) {
+            System.out.println("The file with such name does not exist");
```
Sofasmile, line 26: https://mate-academy.github.io/jv-program-fulltime/01_core/checklist/hw_05.html
Don’t ignore exceptions. Re-throw them as RuntimeException(e.g. throw new RuntimeException(e)) if you don’t expect your code to fail here, or make necessary actions at the catch block to fix the problem.

jlightgreen: done

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,26 @@
...
  */
 public class FileWork {
+    private static final String FIRST_CHARACTER = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder fileContent = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            while (reader.ready()) {
+                fileContent.append(reader.readLine());
```
KseniiaMakarova, line 23: it seems the lines are not separate from each other after you append them to StringBuilder

jlightgreen: done

PR #83, opened 2020-07-31
hw5-3_done

vladkuzya:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,52 @@
...
+                }
+            }
+            String[] result = resultString.trim().split(" ");
+            Arrays.sort(result);
+
+            return result[0] != "" ? result : new String[0];
+        } catch (FileNotFoundException e) {
+            System.out.println("File with name " + fileName + " not found!");
+        } catch (IOException e) {
+            e.printStackTrace();
```
boroda4436, line 48: Don’t ignore exceptions. Re-throw them as RuntimeException(e.g. throw new RuntimeException(e)) if you don’t expect your code to fail here, or make necessary actions at the catch block to fix the problem.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,52 @@
...
+                } else {
+                    return new String[0];
+                }
+            }
+            String[] result = resultString.trim().split(" ");
+            Arrays.sort(result);
+
+            return result[0] != "" ? result : new String[0];
+        } catch (FileNotFoundException e) {
+            System.out.println("File with name " + fileName + " not found!");
```
boroda4436, line 46: same here

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,53 @@
...
-        return null;
+    public static final char LETTER = 'w';
+
+    public static String[] readFromFile(String fileName) {
+        try (FileReader fileReader = new FileReader(fileName)) {
+            Scanner scanner = new Scanner(fileReader);
+
+            String text = "";
+            while (scanner.hasNextLine()) {
+                text += scanner.nextLine() + " ";
```
dmytro-hryhoruk, line 26: let's use StringBuilder and it's method append().
Tell me why that would be better in comment below?

src/main/java/core/basesyntax/FileWork.java
dmytro-hryhoruk: now we can delete that System.out.println

src/main/java/core/basesyntax/FileWork.java
dmytro-hryhoruk: same here about e.printStackTrace();

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,53 @@
...
+            }
+            String[] words = text.split(" ");
+            String resultString = "";
+
+            int index = 0;
+            for (int i = 0; i < words.length; i++) {
+                if (words.length > 1) {
+                    if (words[i].toLowerCase().charAt(0) == LETTER) {
```
dmytro-hryhoruk, line 34: we have string method to check if it's starts with character

src/main/java/core/basesyntax/FileWork.java
vladkuzya: StringBuilder работает намного быстрее String. И если бы у нас был текстовый файл на миллион строк и их нужно было все считать, то это была бы большая разница во времени исполнения между StringBuilder и String

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,51 @@
...
+                text.append(scanner.nextLine() + " ");
+            }
+            String[] words = text.toString().split(" ");
+            String resultString = "";
+
+            int index = 0;
+            for (int i = 0; i < words.length; i++) {
+                if (words.length > 1) {
+                    if (words[i].toLowerCase().startsWith(LETTER)) {
+                        resultString += words[i].toLowerCase().replaceAll("[.), \\-'!?]", "") + " ";
```
dmytro-hryhoruk, line 35: try to find regexp for that

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,51 @@
...
+                }
+            }
+            String[] result = resultString.trim().split(" ");
+            Arrays.sort(result);
+
+            return result[0] != "" ? result : new String[0];
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
dmytro-hryhoruk, line 48: If you are handling these two type of exceptions in the same way. You can just leave catch with IOException

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,51 @@
...
+                        resultString += words[i].toLowerCase().replaceAll("[.), \\-'!?]", "") + " ";
+                    }
+                } else {
+                    return new String[0];
+                }
+            }
+            String[] result = resultString.trim().split(" ");
+            Arrays.sort(result);
+
+            return result[0] != "" ? result : new String[0];
```
dmytro-hryhoruk, line 44: remove this check either here or on 33 line

PR #85, opened 2020-07-31
task done

Rommelua:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,15 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        try {
+            return Files.lines(Path.of(fileName))
+                    .map(String::toLowerCase)
+                    .flatMap(line -> Stream.of(line.split("\\W")))
+                    .filter(word -> word.startsWith("w"))
```
Sofasmile, line 21: let's create constant for "w"

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,15 @@
...
+                    .sorted().toArray(String[]::new);
+        } catch (IOException e) {
+            e.printStackTrace();
```
Sofasmile, line 24: Don’t ignore exceptions. Re-throw them as RuntimeException(e.g. throw new RuntimeException(e)) if you don’t expect your code to fail here, or make necessary actions at the catch block to fix the problem.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +14,15 @@
...
+        }
+        return new String[0];
```
Sofasmile, line 26: remove this after fixing previous comment))

PR #88, opened 2020-07-31
ready solution hw-5.3

RomanHreits:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +17,32 @@
...
+            try {
+                Files.lines(pathToFile)
+                        .map(x -> Arrays.stream(x.split("\\W"))
+                                .map(String::toLowerCase)
+                                .filter(z -> {
+                                    if (z.startsWith("w")) {
+                                        stringsFromFile.add(z);
+                                        return true;
+                                    }
+                                    return false;
```
dmytro-hryhoruk, line 32: it's good that you are using streams already. Im sure that you can simplify that solution and even remove arrayCreator method. 
Hint: try to use stream api methods sorted() and collect()

RomanHreits: done

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +15,16 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        Path pathToFile = Paths.get(fileName);
+        if (Files.exists(pathToFile) & Files.isReadable(pathToFile)) {
+            try {
+                return Arrays.stream(Files.readString(pathToFile)
+                        .split("\\W")).map(String::toLowerCase)
+                        .filter(x -> x.startsWith("w")).sorted().toArray(String[]::new);
```
dmytro-hryhoruk, line 23:
```suggestion
                return Arrays.stream(Files.readString(pathToFile)
                        .split("\\W"))
                        .map(String::toLowerCase)
                        .filter(x -> x.startsWith("w"))
                        .sorted()
                        .toArray(String[]::new);
```
Let's write each stream method in a new line. This way code will be more readable

PR #89, opened 2020-07-31
homework5.3

Alkey:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,40 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+
+    private static final String CONSTANTECHAR = "w";
```
KseniiaMakarova, line 19: let's make the name more clear
something like `FIRST_CHARACTER`

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,40 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder builder = new StringBuilder();
+        try {
+            Path paths = Paths.get(fileName);
+            File file = paths.toFile();
```
KseniiaMakarova, line 25: you can get File from fileName without using Path

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,40 @@
...
+
+            while (inputStream.available() > 0) {
+                char tmp = (char) inputStream.read();
+                builder.append(tmp);
+            }
+            inputStream.close();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        String[] words = builder.toString().toLowerCase().split(" ");
```
KseniiaMakarova, line 36: do words on different lines get separated in this case?
what should we use in `split()` to make this work?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,40 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        String[] lines = builder.toString().toLowerCase().split("\n");
+        builder.delete(0, builder.length());
+        for (String line : lines) {
+            String[] words = line.split(" ");
+            for (String word : words) {
+                if (word.startsWith(FIRST_CHARACTER)) {
+                    builder.append(word).append(" ");
```
KseniiaMakarova, line 38: let's do .replaceAll() here as well

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +13,40 @@
...
+                }
+            }
+        }
+        if (builder.length() == 0) {
+            return new String[]{};
+        }
+        String[] result = builder.toString().trim().split(" ");
+        for (int i = 0; i < result.length; i++) {
+            result[i] = result[i].replaceAll("\\W", "");
+        }
```
KseniiaMakarova, line 48: we can remove this if we move this logic to line 38

PR #93, opened 2020-08-01
added solution

Victoria-Moiseienko:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,24 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    final String constantW = "w";
```
dmytro-hryhoruk, line 17: check [constant naming convention](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html)

Victoria-Moiseienko: Updated. Done

PR #94, opened 2020-08-01
Solution added

Nokafan:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,22 @@
...
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> resultWords = new ArrayList<>();
+        try {
+            List<String> inputLines = Files.readAllLines(Paths.get(fileName));
+            for (String string : inputLines) {
+                for (String oneWord : string.toLowerCase()
+                        .split("[^a-z]+")
+                ) {
+                    if (oneWord.charAt(0) == 'w') {
```
KseniiaMakarova, line 26: let's make `w` a constant field

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,22 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> resultWords = new ArrayList<>();
+        try {
+            List<String> inputLines = Files.readAllLines(Paths.get(fileName));
+            for (String string : inputLines) {
```
KseniiaMakarova, line 22:
```suggestion
            for (String line : inputLines) {
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +16,22 @@
...
+                for (String oneWord : string.toLowerCase()
+                        .split("[^a-z]+")
+                ) {
+                    if (oneWord.charAt(0) == 'w') {
+                        resultWords.add(oneWord);
+                    }
+                }
+            }
+        } catch (IOException e) {
+            return new String[]{};
```
KseniiaMakarova, line 32: let's rethrow a RuntimeException here

src/main/java/core/basesyntax/FileWork.java
Nokafan: Wrong push. I don't have it in my code :(

Nokafan: Question: is it ok to split declaring
 ` List<String> inputLines = null;
        try {
            inputLines = Files.readAllLines(Paths.get(fileName));
        } catch (IOException e) {`
and put in the try block only one readAllLines?

src/main/java/core/basesyntax/FileWork.java
Nokafan: Fixed

PR #95, opened 2020-08-01
solved

ilya449:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,26 @@
...
+    private final String emptyString = "";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        final Path path = Paths.get(fileName);
+        try {
+            if (Files.exists(path)
+                    && Files.isReadable(path)
+                    && Files.size(path) > 0) {
+                return Arrays.stream(Files.readString(path).split(space))
```
KseniiaMakarova, line 28: you are splitting by " " here
do words separated by "\n" get split?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,26 @@
...
  */
 public class FileWork {
+    private final String regex = "\\W";
+    private final String letter = "w";
+    private final String space = " ";
+    private final String emptyString = "";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        final Path path = Paths.get(fileName);
```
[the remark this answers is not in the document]
ilya449: because it doesn't change

src/main/java/core/basesyntax/FileWork.java
ilya449: yes

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,26 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    private final String regex = "\\W";
+    private final String letter = "w";
+    private final String space = " ";
+    private final String emptyString = "";
```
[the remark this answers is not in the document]
ilya449: when I named them in UPPER_CASE I had problems with check style.
but now I've made them static and everything is OK, thank you

src/main/java/core/basesyntax/FileWork.java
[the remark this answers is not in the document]
KseniiaMakarova: I think `final` is redundant here
it is used when we explicitly indicate that we want a variable to never point at any other value

src/main/java/core/basesyntax/FileWork.java
KseniiaMakarova: unfortunately no
`Files.readString(path)` - here you get one big line
and line breaks look the following  - `\n'

when you do `.split(" ") then, the words separated with `\n' remain together

src/main/java/core/basesyntax/FileWork.java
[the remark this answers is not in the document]
ilya449: Ok, I've changed it

src/main/java/core/basesyntax/FileWork.java
ilya449: I've just tried to split string line which contains several ```\n``` and ```" "```  by ```.split(" ")``` and it was splitted everywhere...

ilya449: Do you mean that I need to use ```\\s+``` instead ```" "```? I've found it and changed

PR #96, opened 2020-08-01
Homework 5.3 completed

Borisko130:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,30 @@
...
     public String[] readFromFile(String fileName) {
-        return null;
+        Path path = Paths.get(fileName);
+        List<String> selectedWords = new ArrayList<>();
+
+        try (BufferedReader reader = Files.newBufferedReader(path)) {
+            String buffer;
+            while ((buffer = reader.readLine()) != null) {
+                for (String s : buffer.split(" ")) {
+                    if (s.toLowerCase().charAt(0) == WORDS_START_FROM) {
```
dmytro-hryhoruk, line 31: we have string method to check if string starts with character

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,30 @@
...
+                    }
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("File not found", e);
+        }
+
+        if (selectedWords.size() == 0) {
+            return new String[0];
+        }
```
dmytro-hryhoruk, line 42: do you really need this check?

PR #97, opened 2020-08-01
added solution

AnaLanda:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,46 @@
...
+        List<String> lines = new ArrayList<String>();
+        try {
+            lines = Files.readAllLines(Paths.get(fileName), UTF_8);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        List<String> wordsStartingWithW = new ArrayList<String>();
+        for (String line : lines) {
+            String[] words =
+                    line.replaceAll("[^A-Za-z ]", "").toLowerCase().split(" ");
```
dmytro-hryhoruk, line 35:
```suggestion
                    line.toLowerCase().replaceAll("[^a-z ]", "").split(" ");
```

dmytro-hryhoruk: that doesn't make difference actually, but it's better to keep your regexp as simple as possible

PR #98, opened 2020-08-01
Home work5.3

KostiantynVorobiov:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
 public class FileWork {
+    static final String LETTER = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> rowOfText = null;
+        try {
+            rowOfText = Files.readAllLines(Paths.get(fileName));
+        } catch (IOException e) {
+            throw new RuntimeException();
```
dmytro-hryhoruk, line 26: Let's add exception object and some informative to constructor. Otherwise we won't have any information about what has happened

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
 /**
  * <p>Дано файл, потрібно прочитати його вміст і вибрати всі слова що починаються на `w`.
  * Результат повернути у вигляді відсортованого масиву (за замовчуванням). Всі слова повинні
  * бути в нижньому регістрі. У випадку якщо таких слів не знайдено повернути пустий масив.
  * Приклад: Width world Wide web
  * Результат: web wide width world</p>
  */
+
 public class FileWork {
+    static final String LETTER = "w";
```
dmytro-hryhoruk, line 19: let's also add access modifier

src/main/java/core/basesyntax/FileWork.java
KostiantynVorobiov: done

src/main/java/core/basesyntax/FileWork.java
KostiantynVorobiov: done

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,42 @@
...
+    private static final String LETTER = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> rowOfText = null;
+        try {
+            rowOfText = Files.readAllLines(Paths.get(fileName));
+        } catch (IOException e) {
+            throw new RuntimeException("File not found");
+        }
```
```suggestion
        } catch (IOException e) {
            throw new RuntimeException("File not found", e);
        }
```
that's what i meant) We should save the original content of the exception that has been catched

KostiantynVorobiov: thank you! done.

PR #100, opened 2020-08-02
added solution

tarkaiev:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,29 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    private static final String WORD_STARTS_WITH = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        Path inputFilePath = Paths.get(fileName);
+        ArrayList<String> inputFile = new ArrayList<>();
```
KseniiaMakarova, line 22: in this case a good practice is to use type of a parent
```suggestion
        List<String> inputFile = new ArrayList<>();
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,29 @@
...
+        try {
+            inputFile = (ArrayList<String>) Files.readAllLines(inputFilePath);
```
KseniiaMakarova, line 24: we do not need casting here

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,29 @@
...
+        } catch (IOException e) {
+            System.out.println("File does not exist");
+        }
+        ArrayList<String> result = new ArrayList<>();
+        for (String string : inputFile) {
+            String[] words = string.split(" ");
+
+            for (String word : words) {
+                if (word.replaceAll("\\W", "").toLowerCase().startsWith(WORD_STARTS_WITH)) {
+                    result.add(word.replaceAll("\\W", "").toLowerCase());
```
KseniiaMakarova, line 34: let's first modify the `word` String, and then check if it starts with `WORD_STARTS_WITH`

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,29 @@
...
+                }
+            }
+        }
+        Collections.sort(result);
+
+        return result.toArray((new String[result.size()]));
+
```
KseniiaMakarova, line 41: you can remove this blank line

src/main/java/core/basesyntax/FileWork.java
tarkaiev: fixed

src/main/java/core/basesyntax/FileWork.java
tarkaiev: fixed

src/main/java/core/basesyntax/FileWork.java
tarkaiev: done

src/main/java/core/basesyntax/FileWork.java
tarkaiev: removed

PR #101, opened 2020-08-02
Add solution

OleksiiTromsa:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,29 @@
...
+            throw new RuntimeException("Some problems with file");
+        }
+
+        if (fileText.length() == 0) {
+            return new String[0];
+        }
+
+        return Arrays.stream(fileText.split("\\W+"))
+                .map(String::toLowerCase)
+                .filter(s -> s.substring(0, 1).equals(FIRST_LETTER))
```
KseniiaMakarova, line 25: there is a predefined method that makes this check

PR #103, opened 2020-08-02
Build array with "w"- words from file

fyarkiy:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,24 @@
...
 public class FileWork {
-    public String[] readFromFile(String fileName) {
+    private static final String SMALL_LETTER_W = "w";
+
+    public String[] readFromFile(String fileName) throws RuntimeException {
+        try {
+            List<String> sortedWordsWithW = new ArrayList<>();
+            for (String line : Files.readAllLines(Paths.get(fileName))) {
+                for (String word : line.toLowerCase().split(" ")) {
+                    if (word.substring(0, 1).equals(SMALL_LETTER_W)) {
```
KseniiaMakarova, line 25: there is a predefined method in class String for this purpose

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,24 @@
...
+                        sortedWordsWithW.add(word.replaceAll("[^a-z\\-]+",""));
```
KseniiaMakarova, line 26:
```suggestion
                        sortedWordsWithW.add(word.replaceAll("\\W+",""));
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,24 @@
...
+                    }
+                }
+            }
+            Collections.sort(sortedWordsWithW);
+            String[] result = sortedWordsWithW.toArray(new String[0]);
+            return result;
+        } catch (IOException e) {
+            e.printStackTrace();
```
KseniiaMakarova, line 34: let's rethrow this as a RuntimeException

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,24 @@
...
+        }
         return null;
```
KseniiaMakarova, line 36: remove this line after fixing the previous issue

src/main/java/core/basesyntax/FileWork.java
fyarkiy: Do you mean matches instead of equals? Then I don't need constant - some improvement in code. But if something else, pls let me know. Thanks!!!

KseniiaMakarova: let's use `word.startsWith(String prefix)` method

KseniiaMakarova: bring back your constant and use it as a prefix

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,21 @@
...
+                    if (word.substring(0, 1).matches("w")) {
+                        sortedWordsWithW.add(word.replaceAll("\\W+",""));
+                    }
+                }
+            }
+            Collections.sort(sortedWordsWithW);
+            String[] result = sortedWordsWithW.toArray(new String[0]);
+            return result;
+        } catch (IOException e) {
+            throw new RuntimeException();
```
KseniiaMakarova, line 32: any exception should be informative, right?
```suggestion
            throw new RuntimeException(e);
```

src/main/java/core/basesyntax/FileWork.java
fyarkiy: sure = on this case I need constant. But matches looks cleaner:)

src/main/java/core/basesyntax/FileWork.java
dmytro-hryhoruk: @fyarkiy fix this comment too

PR #104, opened 2020-08-02
HOMEWORK-5-3: - done;

jah701:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,26 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    private final String letter = "w";
```
dmytro-hryhoruk, line 18: check [constant naming convention](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html)

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +15,26 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+    private final String LETTER = "w";
```
dmytro-hryhoruk, line 18:
```suggestion
    private static final String LETTER = "w";
```

PR #108, opened 2020-08-02
Added solution

AndrewPolishchuc:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +16,29 @@
...
+    public static String[] readFromFile(String fileName) {
+        String sampleLine;
+        File sampleFile = new File(fileName);
+        ArrayList<String> answerList = new ArrayList<>();
+        try (BufferedReader sampleBuffer = new BufferedReader(new FileReader(sampleFile))) {
+            while ((sampleLine = sampleBuffer.readLine()) != null) {
+                String[] words = sampleLine.split(" ");
+                for (String word : words) {
+                    if (word.toLowerCase().charAt(0) == SEARCHLETTER) {
+                        answerList.add(word.toLowerCase().replaceAll("[^A-Za-zА-Яа-я0-9]",""));
```
KseniiaMakarova, line 30:
```suggestion
                        answerList.add(word.toLowerCase().replaceAll("\\W+",""));
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +16,29 @@
...
+                    }
+                }
+            }
+        } catch (FileNotFoundException e) {
```
KseniiaMakarova, line 34: try catching both Exceptions in the same catch block using pipe `|`

src/main/java/core/basesyntax/FileWork.java
AndrewPolishchuc: Done

src/main/java/core/basesyntax/FileWork.java
AndrewPolishchuc: We can`t use | for catching both Exceptions. Types in multi-catch must be disjoint: 'java.io.FileNotFoundException' is a subclass of 'java.io.IOException'. We can only catch java.io.IOException.

PR #109, opened 2020-08-02
added solution

aleksandr-hrankin:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,28 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                matcher = pattern.matcher(line.toLowerCase());
+                while (matcher.find()) {
+                    wordList.add(matcher.group());
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("File not found!");
```
dmytro-hryhoruk, line 36: let's also pass exception object to constructor(overloaded) in order not to lose information about what has happened

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +17,28 @@
...
+        }
+
+        String[] words = wordList.toArray(new String[0]);
+        Arrays.sort(words);
+
```
dmytro-hryhoruk, line 41:
```suggestion
```

PR #112, opened 2020-08-02
solution added

margarita-ignatieva:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,34 @@
...
- */
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 public class FileWork {
+    static final String wLetter = "w";
```
dmytro-hryhoruk, line 11: check constant [naming convention](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html)

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,34 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> resultList = new ArrayList<>();
+        List<String> lines = new ArrayList<>();
+        try {
+            lines = Files.readAllLines(Paths.get(fileName));
+
+            for (String line: lines) {
+                String[] words = line.toLowerCase().replaceAll("[^a-z]"," ").split(" ");
```
dmytro-hryhoruk, line 20:
```suggestion
                String[] words = line.toLowerCase().split("\\W+");
```
Try this way. But write an explanation above this comment what \\W+ regex does)

src/main/java/core/basesyntax/FileWork.java
margarita-ignatieva: done

src/main/java/core/basesyntax/FileWork.java
margarita-ignatieva: done

PR #114, opened 2020-08-02
solution added

Vosker:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,25 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Arrays;
+
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+    private static final String WORDS_START = "w";
+
+    public static String[] readFromFile(String fileName) {
```
dmytro-hryhoruk, line 12: why static?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,25 @@
...
+        try {
+            return Arrays.stream(Files.readString(Paths.get(fileName), UTF_8)
+                    .toLowerCase()
+                    .split("\\W"))
+                    .filter(x -> x.startsWith(WORDS_START))
+                    .sorted()
+                    .toArray(String[]::new);
+        } catch (Exception e) {
+            return null;
```
dmytro-hryhoruk, line 21: let's rethrow exception

PR #116, opened 2020-08-03
hw

YaTut1901:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +15,25 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        final char W = 'w';
+        ArrayList<String> wordsWithWList = new ArrayList<String>();
+        String text = "";
+        try {
+            text = new String(Files.readAllBytes((Paths.get(fileName)))).toLowerCase().replaceAll(
+                    "[^A-Za-zА-Яа-я ]", "");
```
KseniiaMakarova, line 23: your solution has one drawback:
the last word on the line is sticked together with the first word on the next line

`word1 word2
word 3` 
=> `word1 word2word3`

let's try retrieving lines first, and then split each line

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -9,6 +15,25 @@
...
+        } catch (IOException e) {
+            System.out.println("Такого файла не существует");
```
KseniiaMakarova, line 25: we should rethrow an informative RuntimeException here

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -15,21 +15,23 @@
  */
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        final char W = 'w';
+        final char CONSTW = 'w';
```
KseniiaMakarova, line 18: add public and static keywords

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -14,8 +14,10 @@
  * Результат: web wide width world</p>
  */
 public class FileWork {
+
+    public static final char CONSTW = 'w';
```
dmytro-hryhoruk, line 18:
```suggestion
    public static final char CONSTANT_W = 'w';
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,30 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+        ArrayList<String> wordsWithWList = new ArrayList<String>();
+        String text = "";
+        try {
+            text = new String(Files.readAllBytes((Paths.get(fileName)))).toLowerCase().replaceAll(
```
dmytro-hryhoruk, line 24: let's use readAllLines() method. This way you will get List<String> and won't have to use split("\n") to divide text to lines

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +14,30 @@
...
+                    "[^A-Za-zА-Яа-я\n ]", "");
+        } catch (IOException e) {
+            throw new RuntimeException("Такого файла не существует", e);
```
dmytro-hryhoruk, line 27: let's use english language in our programs)Anybody in the world should be able to understand that

PR #120, opened 2020-08-03
hw 5-3

mikhail771:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +16,37 @@
...
+            while (line != null) {
+                result.append(line);
+                line = br.readLine();
+            }
+            words = result.toString().split(" ");
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("Can't read file", e);
+        } catch (IOException exception) {
+            exception.printStackTrace();
+        }
```
dmytro-hryhoruk, line 37: do you need two catch block here?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -8,7 +16,37 @@
...
+
+        for (String word : words) {
+            word = word.toLowerCase().replaceAll("[^A-Za-z]", "");
+            if (word.indexOf(SEARCH_LETTER) == 0) {
```
dmytro-hryhoruk, line 41: we have string method to check if it's starts with some character

dmytro-hryhoruk, line 41: let's make this manipulation while reading files. This way you will get rid of using ArrayList

src/main/java/core/basesyntax/FileWork.java
mikhail771: reedLine() method requires second exception

src/main/java/core/basesyntax/FileWork.java
mikhail771: done

PR #125, opened 2020-08-03
HW-5.3

vremere-a:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,47 @@
...
     public String[] readFromFile(String fileName) {
-        return null;
+        String tempString = EMPTY;
+        String[] result;
+        try (BufferedReader newReader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = newReader.readLine()) != null) {
+                String[] arr1 = line.toLowerCase().split(SPACE);
+                for (String line2 : arr1) {
+                    if (line2.substring(0, 1).equals(LETTER)) {
```
dmytro-hryhoruk, line 31: see if string have a special method to check whether the first character is equal to some

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,47 @@
...
+                        tempString += line2.replace(POINT, EMPTY).replace(QUESTION_MARK, EMPTY)
+                                .replace(EXCLAMATION_MARK, EMPTY) + SPACE;
```
dmytro-hryhoruk, line 33: try to find regular expression for that.

src/main/java/core/basesyntax/FileWork.java
vremere-a: done

src/main/java/core/basesyntax/FileWork.java
vremere-a: done

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,46 @@
...
  * Результат: web wide width world</p>
  */
+
 public class FileWork {
+    static final String LETTER = "w";
+    static final String POINT = ".";
+    static final String EMPTY = "";
+    static final String SPACE = " ";
+    static final String EXCLAMATION_MARK = "!";
+    static final String QUESTION_MARK = "?";
```
dmytro-hryhoruk, line 21: 1) add access modifiers to your constants;
2) Do you need all of them?

vremere-a: done

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,43 @@
...
-        return null;
+        String tempString = EMPTY;
+        String[] result;
+        try (BufferedReader newReader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = newReader.readLine()) != null) {
+                String[] arr1 = line.toLowerCase().split(SPACE);
+                for (String line2 : arr1) {
+                    if (line2.startsWith(LETTER)) {
+                        tempString += line2.replaceAll("[^a-z0-9]", "") + SPACE;
```
dmytro-hryhoruk, line 29: let's use String builder here. Tell me why this would be a better solution in comment below

vremere-a: done. 
StringBuilder have more economy than String. 
and can expand

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,44 @@
...
-        return null;
+        StringBuilder tempString = new StringBuilder();
+        String[] result;
+        try (BufferedReader newReader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = newReader.readLine()) != null) {
+                String[] arr1 = line.toLowerCase().split(SPACE);
+                for (String line2 : arr1) {
+                    if (line2.startsWith(LETTER)) {
+                        tempString.append(line2.replaceAll("[^a-z0-9]", "") + SPACE);
```
dmytro-hryhoruk, line 29:
```suggestion
                        tempString.append(line2.replaceAll("[^a-z0-9]", "")).append(SPACE);
```

PR #126, opened 2020-08-03
hw-5(3)

glones:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,31 @@
...
+import java.util.List;
+
 public class FileWork {
+    private static final String LETTER = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        try {
+            List<String> preResult = new ArrayList<String>();
+            String file = Files.readString(Paths.get(fileName)).toLowerCase();
```
KseniiaMakarova, line 16:
```suggestion
            String text = Files.readString(Paths.get(fileName)).toLowerCase();
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,31 @@
...
+            if (file.length() > 0) {
+                for (String word : file.split("[^A-Za-zА-Яа-я]+")) {
```
KseniiaMakarova, line 18:
```suggestion
                for (String word : file.split("\\W+")) {
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,14 +1,31 @@
...
+                    if (word.substring(0, 1).equals(LETTER)) {
```
KseniiaMakarova, line 19: there is a method in class String that checks if a String starts with a certain value

src/main/java/core/basesyntax/FileWork.java
glones: done

src/main/java/core/basesyntax/FileWork.java
glones: done

src/main/java/core/basesyntax/FileWork.java
glones: done

PR #135, opened 2020-12-03
Read and Filter Task

IvanLuchkin:
Task completed.

### DIFF
--- a/src/main/java/core/basesyntax/FileWork.java
+++ b/src/main/java/core/basesyntax/FileWork.java
@@ -1,7 +1,23 @@
 package core.basesyntax;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Arrays;
+
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+    private static final String START_SYMBOL = "w";
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,23 @@
...
 public class FileWork {
-    public String[] readFromFile(String fileName) {
-        return null;
+    private static final String START_SYMBOL = "w";
+
+    public static String[] readFromFile(String fileName) {
+        try {
+            return Files.readAllLines(new File(fileName).toPath())
+                    .stream()
+                    .flatMap(line -> Arrays.stream((line + " ").toLowerCase().split("[.!?]* ")))
```
Sofasmile, line 15: - why do we use (line + " ") here?
- try to replace "[.!?]* " with "\\W+"

Sofasmile (changes requested):

src/main/java/core/basesyntax/FileWork.java
IvanLuchkin: If we don't use "\\W+" we have to add an extra space so that the dot in the end of the file is not attached to the word after parsing

dmytro-hryhoruk (approved):
Good job

PR #165, opened 2020-12-06
1) my solution

AntonPanchishko:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,42 @@
...
+        String[] resultW = new String[0];
+        StringBuilder stringBuilder = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
+            String value = reader.readLine();
+            if (value == null) {
+                return resultW;
+            }
+            reader.read();
+            while (value != null) {
```
dmytro-hryhoruk, line 20: it's duplication checks, think how we can remove them and leave only while the condition

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,42 @@
...
+                stringBuilder.append(value).append(" ");
+                value = reader.readLine();
+            }
+            String[] split = stringBuilder.toString().toLowerCase().split(" ");
+            StringBuilder stringBuilder1 = new StringBuilder();
```
dmytro-hryhoruk, line 25: don't use numbers in naming

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,42 @@
...
+            for (int i = 0; i < split.length; i++) {
+                split[i] = split[i].replaceAll("[.!?]", "");
+                if (split[i].charAt(0) == 'w') {
+                    stringBuilder1 = stringBuilder1.append(split[i]).append(" ");
+                }
+            }
+            resultW = stringBuilder1.toString().split(" ");
+            if (stringBuilder1.toString().equals("")) {
```
dmytro-hryhoruk, line 33: let's check the length of stringbuilder

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,42 @@
...
+                return new String[0];
+            }
+            Arrays.sort(resultW);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file");
```
dmytro-hryhoruk, line 38: What file?Let's add the original exception to the overloaded constructor too

dmytro-hryhoruk (changes requested):

PR #168, opened 2020-12-06
1) Commit

KostyantynProskura:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+            throw new RuntimeException("Can't read file " + fileName, e);
+        }
+    }
+
+    private String[] wordsFromListLowerCase(List<String> listOfWords) {
+        StringBuilder textBuilder = new StringBuilder();
+
+        for (String word : listOfWords) {
+            textBuilder.append(word.toLowerCase()).append(System.lineSeparator());
+            if (textBuilder.length() <= 0) {
```
```suggestion
            if (textBuilder.length() == 0) {
```
It never will be less than zero

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        textBuilder = new StringBuilder();
+
+        for (String word : wordsArray) {
+            if (startWithLetter(word)) {
+                textBuilder.append(word).append(" ");
+            }
+        }
+        wordsArray = textBuilder.toString().split(" ");
+        Arrays.sort(wordsArray);
+        return wordsArray.length == 1 ? new String[0] : wordsArray;
```
dmytro-hryhoruk, line 42: let's just check if the builder has length zero then return empty array

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,51 @@
...
+            if (textBuilder.length() == 0) {
+                return new String[0];
+            }
+        }
+
+        String[] wordsArray = textBuilder.toString().toLowerCase().split("[^\\w]");
+        textBuilder = new StringBuilder();
+
+        for (String word : wordsArray) {
+            if (startWithLetter(word)) {
```
dmytro-hryhoruk, line 36: don't see the point to wrap only one string method(startsWith() ) with a private one

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,51 @@
...
+
 public class FileWork {
+    private static final String SPECIFIED_CHARACTER = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        File file = new File(fileName);
+        try {
+            List<String> listOfWords = Files.readAllLines(new File(fileName).toPath());
+            return wordsFromListLowerCase(listOfWords);
```
dmytro-hryhoruk, line 16: let's call this method after the catch block

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,51 @@
...
+        }
+        if (textBuilder.toString().length() == 0) {
+            return new String[0];
+        }
+        wordsArray = textBuilder.toString().split(" ");
+        Arrays.sort(wordsArray);
+        return wordsArray.length == 0 ? new String[0] : wordsArray;
+    }
+
+    public boolean startWithLetter(String word) {
```
dmytro-hryhoruk, line 48: why public?

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+
+        String[] wordsArray = textBuilder.toString().toLowerCase().split("[^\\w]");
+        textBuilder = new StringBuilder();
+
+        for (String word : wordsArray) {
+            if (word.startsWith(SPECIFIED_CHARACTER)) {
+                textBuilder.append(word).append(" ");
+            }
+        }
+        if (textBuilder.toString().length() == 0) {
```
dmytro-hryhoruk, line 41:
```suggestion
        if (textBuilder.toString().isEmpty()) {
```

dmytro-hryhoruk (approved):

PR #170, opened 2020-12-06
Initial

Max-Pochepets:


dmytro-hryhoruk (changes requested):
You are doing the same job twice in your solution(reading file, checking on first symbol 'W');
We definitely can simplify that.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,41 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Locale;
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            StringBuilder builder = new StringBuilder();
+            int value = bufferedReader.read();
```
dmytro-hryhoruk, line 14: let's use readLine() method. Then you can split each line with regexp = \\W+ and check whether or not it starts with letter using only one check: Character.toLowerCase(word.chartAt(0)) == 'w'

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #175, opened 2020-12-07
created logic, that reads from file and returns result

AndriiBardin:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,39 @@
...
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String[] empty = new String[]{};
+        StringBuilder readStrings = new StringBuilder();
+        File test = new File(fileName);
+
+        if (test.length() == 0) {
+            return empty;
+        } else {
```
KseniiaMakarova, line 18: do we need to do `else` after return?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,39 @@
...
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String[] empty = new String[]{};
+        StringBuilder readStrings = new StringBuilder();
+        File test = new File(fileName);
+
+        if (test.length() == 0) {
+            return empty;
```
KseniiaMakarova, line 17: what do you do with this check?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,39 @@
...
+        } else {
+            try {
+                List<String> lines = Files.readAllLines(Path.of(fileName));
+                String[] text = lines.toArray(new String[0]);
+                for (String line : text) {
+                    String[] temp = line.split(" ");
+                    for (String word : temp) {
+                        if (word.toLowerCase().startsWith("w")) {
+                            readStrings.append(word).append(",");
```
KseniiaMakarova, line 26: let's move `w` and the delimiter to constant fields

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,39 @@
...
+                        }
+                    }
+                }
+            } catch (IOException e) {
+                throw new RuntimeException("Can't read file");
```
KseniiaMakarova, line 31:
```suggestion
                throw new RuntimeException("Can't read file", e);
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/FileWork.java
AndriiBardin: was checking if file is null but now, it is not necessary, doing the same check in the return

KseniiaMakarova (approved):

PR #177, opened 2020-12-07
solved problem

Sashk52:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+
 public class FileWork {
+    private static final String FIRST_SYMBOL = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        BufferedReader bufferedReader = null;
```
dmytro-hryhoruk, line 12: let's declare and initialize the reader in try with resources

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+            throw new RuntimeException("Can't read from file", e);
+        } finally {
+            if (bufferedReader != null) {
+                try {
+                    bufferedReader.close();
+                } catch (IOException e) {
+                    throw new RuntimeException("Can't close BufferedReader",e);
+                }
+            }
+        }
```
dmytro-hryhoruk, line 32: use try with resources

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        BufferedReader bufferedReader = null;
+        StringBuilder stringBuilder = null;
+        try {
+            bufferedReader = new BufferedReader(new FileReader(fileName));
+            stringBuilder = new StringBuilder();
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
```
dmytro-hryhoruk, line 21:
```suggestion
            String value;
            while (value  = bufferedReader.readLine() != null) {
                stringBuilder.append(value).append(System.lineSeparator());
            }
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        } finally {
+            if (bufferedReader != null) {
+                try {
+                    bufferedReader.close();
+                } catch (IOException e) {
+                    throw new RuntimeException("Can't close BufferedReader",e);
+                }
+            }
+        }
+        String[] resultArray = stringBuilder.toString().toLowerCase().split("\\s+");
```
dmytro-hryhoruk, line 33:
```suggestion
        String[] resultArray = stringBuilder.toString().toLowerCase().split("\\W+");
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        StringBuilder newSB = new StringBuilder();
+        for (String res:resultArray) {
+            if (res.startsWith(FIRST_SYMBOL)) {
+                res = res.replaceAll("[!?.,]", "");
```
dmytro-hryhoruk, line 37:
```suggestion
```

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/FileWork.java
Sashk52: done

src/main/java/core/basesyntax/FileWork.java
Sashk52: done

src/main/java/core/basesyntax/FileWork.java
Sashk52: done

src/main/java/core/basesyntax/FileWork.java
Sashk52: Lead to compile error. Should i find solution with this or just ignore this? Compiler require change type of "value" to boolean.

src/main/java/core/basesyntax/FileWork.java
Sashk52: done

dmytro-hryhoruk (approved):

PR #181, opened 2020-12-07
HW5 read from fale

tbondare:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
 
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String [] letterWWords = new String [0];
```
KseniiaMakarova, line 10: let's get used to write like this
```suggestion
        String[] letterWWords = new String[0];
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        StringBuilder buffer = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
+            String line = reader.readLine();
+            String [] wordsOfLine;
```
KseniiaMakarova, line 15: same here
fix this everywhere needed please

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        String [] letterWWords = new String [0];
+        StringBuilder buffer = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
```
KseniiaMakarova, line 13: is this reader closed after usage?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+            String line = reader.readLine();
+            String [] wordsOfLine;
+            while (line != null) {
+                wordsOfLine = line.toLowerCase().split(" ");
+                for (String word : wordsOfLine) {
+                    if (word.charAt(0) == 'w') {
```
KseniiaMakarova, line 19: `w` should be made a constant field - see checklist

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+                        buffer.append(word).append(" ");
+                    }
+                }
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
+            System.out.println("File can't reader");
```
KseniiaMakarova, line 26: let's throw the RuntimeException here with an informative message
and add the original exception to the constructor as the cause

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        for (String word : arrWords) {
+            for (int i = 0; i < word.length(); i++) {
+                if (word.charAt(i) >= 'a' && word.charAt(i) <= 'z') {
+                    string += word.charAt(i);
+                } else {
+                    break;
+                }
+            }
+            string += " ";
+        }
```
KseniiaMakarova, line 43: is it done to remove punctuation? if so, let's use regex and replaceAll() when we filter the right words (line 19)

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
approved to unlock the next topics on the MA platform

PR #184, opened 2021-03-15
Solution

SakVladyslav:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,47 @@
...
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        File file = new File(fileName);
+        StringBuilder stringBuilder = new StringBuilder();
+        String[] resultList = new String[]{};
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
RomanDubovskyi, line 16: it's better to use try with resources instead of try catch finally. This way connection to file will be closed automatically when you exit try block

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,47 @@
...
+            String readerValue = reader.readLine();
+            String regex = "[?.:!;]";
+
+            if (readerValue == null) {
+                return new String[]{};
+            }
+
+            while (readerValue != null) {
+                String[] readerValueList = readerValue.split(" ");
```
RomanDubovskyi, line 25:
```suggestion
            String regex = "[?.:!;]";
            if (readerValue == null) {
                return new String[]{};
            }
            while (readerValue != null) {
                String[] readerValueList = readerValue.split(" ");
```
there is no need to put these empty lines, they make code unnecessary longer

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,47 @@
...
+        File file = new File(fileName);
+        StringBuilder stringBuilder = new StringBuilder();
+        String[] resultList = new String[]{};
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            String readerValue = reader.readLine();
+            String regex = "[?.:!;]";
+
+            if (readerValue == null) {
```
RomanDubovskyi, line 20: you don't need this if condition. This case is already covered by while loop, you won't enter if readerValue is null.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,47 @@
...
+                return new String[]{};
+            }
+
+            while (readerValue != null) {
+                String[] readerValueList = readerValue.split(" ");
+
+                for (String value : readerValueList) {
+                    if (value.toLowerCase().startsWith("w")) {
```
RomanDubovskyi, line 28: declare 'w' as constant, now it's "magic string"

RomanDubovskyi (changes requested):
nice work!

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,43 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+
 public class FileWork {
+    private static final String WORD_FIRST_CHARACTER = "w";
```
RomanDubovskyi, line 10:
```suggestion
    private static final String TARGET_WORDS_START = "w";
```

RomanDubovskyi (approved):

PR #191, opened 2021-04-01
"added solution"

KseniaObertii:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,32 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+
 public class FileWork {
+    private static final char SEARCHED_LETTER = 'w';
+    private static final String SEPARATOR = " ";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        ArrayList<String> result = new ArrayList<>();
```
Resci, line 14: Use abstraction instead implementation where it possible
```suggestion
        List<String> result = new ArrayList<>();
```

Resci (changes requested):

Resci (approved):

PR #205, opened 2021-04-03
Hw5.1

Vlad-Andr:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,61 @@
...
+
 public class FileWork {
+    private static final String SPECIFIED_CHARACTER = "w";
+    private static final String SPACE = "";
+    private static final String SPECIAL_CHARACTERS = "[,\\s\\-:\\?]";
+    private static final String ONLY_WORD_PATTERN = "[^a-z A-Z]";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder stringBuilder = new StringBuilder();
```
RomanDubovskyi, line 18: let's try to simplify this solution - you don't need 3 loops, now for inside of while is enough. 
Inside while split your line with correct regex and and append lower cased version to the result.
After that you will just need to sort your words.

RomanDubovskyi, line 18: name your variables according to their purpose stringBuilder -> filteredWords or smt like this

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,61 @@
...
+
+        Arrays.sort(startWithW);
+        return startWithW;
+    }
+
+    private boolean startWithLetter(String word) {
+        return word.startsWith(SPECIFIED_CHARACTER);
+    }
+
+    private static String getOnlyWord(String str) {
```
RomanDubovskyi, line 56: we don't need these methods if we simplify the logic. 
P.S.
why is it static?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,61 @@
...
+                value = bufferedReader.readLine();
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("File not found", e);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file", e);
+        }
+
+        String[] array = stringBuilder.toString().split(SPECIAL_CHARACTERS);
+        int k = 0;
```
RomanDubovskyi, line 32: k - we could use more informative naming here

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,51 @@
...
+                startWithW[i] = word;
+                i++;
+            }
+        }
+
+        Arrays.sort(startWithW);
+        return startWithW;
+    }
+
+    private boolean startWithLetter(String word) {
```
RomanDubovskyi, line 48: it's not worth to move single line to separate method if it's readable

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,51 @@
...
+    private static final String SPECIAL_CHARACTERS = "[,.\\s\\-:?!]";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder filteredWords = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                filteredWords.append(value.toLowerCase()).append(System.lineSeparator());
+                value = bufferedReader.readLine();
```
RomanDubovskyi, line 19:
```suggestion
            String line = bufferedReader.readLine();
            while (line != null) {
               String[] words = line.toLowerCase().split("\\W+");
                for (String word : words) {
                    if (word.startsWith(SPECIFIED_CHARACTER)) {
                        targetWords.append(word).append(" ");
                    }
                line = bufferedReader.readLine();
```
this way you won't need two loops below, you can split strings by space and sort them

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,41 @@
...
+    private static final String SPACE = " ";
+    private static final String[] EMPTY_ARRAY = new String[0];
+
     public String[] readFromFile(String fileName) {
-        return null;
+        StringBuilder filteredWords = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                String[] words = line.toLowerCase().split("\\W+");
```
RomanDubovskyi, line 19: regex is better to move to constant, its' not as self explainable as empty arr

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,41 @@
...
+                line = bufferedReader.readLine();
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("File not found", e);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file", e);
+        }
+
+        if (filteredWords.length() > 0) {
+            String[] array = filteredWords.toString().split(SPACE);
```
RomanDubovskyi, line 34:
```suggestion
            String[] array = filteredWords.toString().split(" ");
```
your code should be consistent - if you don't use constant in your for loop, then don't use it know.

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,41 @@
...
+            Arrays.sort(array);
+            return array;
+        } else {
+            return EMPTY_ARRAY;
```
RomanDubovskyi, line 38:
```suggestion
            return array;
        } 
            return EMPTY_ARRAY;
```
we don't need else block after return in if

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #218, opened 2021-04-04
home-work done

Aalderaan:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,50 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file...");
+        }
+
+        int count = 0;
```
RomanDubovskyi, line 28: int cout  is redundant, add words to string builder or arraylist and than you will have its size to know if there are any words that match the requirements

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,50 @@
...
+        String[] arrayOfWords = stringBuilder.toString().toLowerCase()
+                .replaceAll("[^a-z ^A-Z]", "").split(" ");
```
RomanDubovskyi, line 30: is you split with the correct regex replaceAll is redundant

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,50 @@
...
+
     public String[] readFromFile(String fileName) {
-        return null;
+
+        File file = new File(fileName);
+        StringBuilder stringBuilder = new StringBuilder();
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            String value = reader.readLine();
```
RomanDubovskyi, line 19:
```suggestion
            String line = reader.readLine();
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,50 @@
...
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
```
RomanDubovskyi, line 21: you can split the line inside of the loop and filter words here, so you will need only to sort words outside

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,32 @@
...
+    private static final char SPECIFIED_CHARACTER = 'w';
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> listOfWords = new ArrayList<>();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
+            String line = reader.readLine();
+            while (line != null) {
+                for (String word : line.split(" ")) {
```
RomanDubovskyi, line 19:
```suggestion
                for (String word : line.toLowerCase().split("\\W+")) {
```
this way you can avoid calling toLowercase twice and  replaceAll()

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,32 @@
...
+                    if (word.toLowerCase().charAt(0) == SPECIFIED_CHARACTER) {
```
RomanDubovskyi, line 20: let's use startsWith

RomanDubovskyi (changes requested):
this mistakes is not fixed yet, close your resources when reading from files 
image

src/main/java/core/basesyntax/FileWork.java
Aalderaan: Roman, Do I need to put finally and close() even if try with resources is autocloseable?

Aalderaan: Sorry, it wasn't "try with recourses"... Now it is.

RomanDubovskyi: if you use try with resourcec you don't need to close resources in finally, that's the beauty of it 🙂

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,31 @@
...
 public class FileWork {
+    private static final String SPECIFIED_CHARACTER = "w";
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> listOfWords = new ArrayList<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line = reader.readLine();
+            while (line != null) {
+                for (String word : line.toLowerCase().split("\\W+")) {
```
RomanDubovskyi, line 18: Ideally it's better to declare regex as constant as well

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,31 @@
...
+                    if (word.startsWith(SPECIFIED_CHARACTER)) {
+                        listOfWords.add(word.replaceAll("\\p{P}", ""));
+                    }
+                }
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file...");
```
RomanDubovskyi, line 26:
```suggestion
            throw new RuntimeException("Can't read the file...", e);
```

RomanDubovskyi (approved):

PR #236, opened 2021-04-05
Hw 004

Koziarevskyi:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            StringBuilder stringBuilder = new StringBuilder();
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
+            data = stringBuilder.toString();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file!", e);
```
Max-Pochepets, line 20: which file?

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        }
+        data = data.replaceAll("[.,?!]", "").trim();
```
Max-Pochepets, line 22: let's make regex a constant variable

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+
```
Max-Pochepets, line 23: remove empty lines from code

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+        if (data.isEmpty()) {
+            String[] emptyArray = new String[0];
+            return emptyArray;
+        }
```
Max-Pochepets, line 27: no need for this check

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+
+        String[] words = data.split("\\s+");
+        int countOfWordsWithW = 0;
+        for (String word : words) {
+            if (word.charAt(0) == 'w' || word.charAt(0) == 'W') {
+                countOfWordsWithW++;
+            }
+        }
```
Max-Pochepets, line 35: you don't have to count valid words, think about StringBuilder

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,48 @@
...
+
+        if (data.isEmpty()) {
+            String[] emptyArray = new String[0];
+            return emptyArray;
+        }
+
+        String[] words = data.split("\\s+");
+        int countOfWordsWithW = 0;
+        for (String word : words) {
+            if (word.charAt(0) == 'w' || word.charAt(0) == 'W') {
```
Max-Pochepets, line 32: it's better to make desired character also a constant

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/FileWork.java
Max-Pochepets: Ok, I see it know, this is good, you can leave it as it is 👍

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -6,42 +6,31 @@
...
-        for (String word : words) {
-            if (word.charAt(0) == 'w' || word.charAt(0) == 'W') {
-                resultArray[i] = word.toLowerCase();
-                i++;
-            }
+        String words = filteredWords.toString();
+        if (words.isEmpty()) {
+            return new String[0];
         }
+        String[] resultArray = filteredWords.toString().split(REGEX_WHITESPACES);
```
Max-Pochepets, line 33: you either use `words` here, or don't create it in the first place. if can manage logic without additional variables - better to do so

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -6,42 +6,31 @@
...
-            String value = bufferedReader.readLine();
-            while (value != null) {
-                stringBuilder.append(value).append(System.lineSeparator());
-                value = bufferedReader.readLine();
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                String[] words = line.toLowerCase().split(REGEX_SYMBOL);
+                for (String word : words) {
+                    if (word.startsWith(LOWER_CASE_W)) {
+                        filteredWords.append(word).append(" ");
```
Max-Pochepets, line 21: every value like this should be constant. this will make it easier if you need to change it. imagine you have 1000 lines of code with this `" "`

Max-Pochepets (approved):

PR #238, opened 2021-04-05
added solution

haniiev:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> result = new ArrayList<>();
+
+        try {
```
RomanDubovskyi, line 14: checklist point about closing resoureces

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> result = new ArrayList<>();
```
RomanDubovskyi, line 12:
```suggestion
        List<String> filteredWords = new ArrayList<>();
```

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
+            StringBuilder stringBuilder = new StringBuilder();
```
RomanDubovskyi, line 16: do we use builder in the code? 
it looks like List is enough

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+
+                for (String word : value.split(" ")) {
```
RomanDubovskyi, line 20:
```suggestion
                for (String word : value.toLowerCase().split("\\W+")) {
```
this way you you won't need to cast each word to lowercase and call replaceAll

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
+                    if (word.toLowerCase().charAt(0) == 'w') {
```
RomanDubovskyi, line 21:
```suggestion
                    if (word.startsWith(// make w constant and compare here)) {
```
startsWith will be better in this case, if in future we will to use combination of chars, etc

RomanDubovskyi (changes requested):
nice solution structure, but we can improve few details

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
-        return null;
+        List<String> filteredWords = new ArrayList<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+
+            String value = reader.readLine();
+            while (value != null) {
+
+                for (String word : value.toLowerCase().split("\\W+")) {
+                    if (word.startsWith(String.valueOf(CONDITIONAL_LATTER))) {
+                        filteredWords.add(word.replaceAll("\\p{P}", ""));
```
RomanDubovskyi, line 22:
```suggestion
                        filteredWords.add(word);
```
we don't need it now

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,7 +1,35 @@
...
+
 public class FileWork {
+    private static final char CONDITIONAL_LATTER = 'w';
+
     public String[] readFromFile(String fileName) {
-        return null;
+        List<String> filteredWords = new ArrayList<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+
+            String value = reader.readLine();
```
RomanDubovskyi, line 17:
```suggestion
            String line = reader.readLine();
```

RomanDubovskyi (approved):

PR #714, opened 2023-02-25
solution of the task

Anestheziv:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,42 @@
...
 public class FileWork {
+    private static final String DELIMITER = "\\W+";
+    private static final String DELIMITER_FOR_RESULT = " ";
+    private static final char LETTER = 'w';
+
     public String[] readFromFile(String fileName) {
-        //write your code here
-        return null;
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
```
olekskov, line 15: Please use try-with-resources for classes that implement AutoClosable interface

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,42 @@
...
+            if (value == null) {
+                return new String[] {};
+            }
+            while (value != null) {
+                stringBuilder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
+            String[] split = stringBuilder.toString().toLowerCase().split(DELIMITER);
+            Arrays.sort(split);
+            StringBuilder builder = new StringBuilder();
```
olekskov, line 27: please choose name that will indicate content of StringBuilder

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,42 @@
...
+    private static final String DELIMITER = "\\W+";
+    private static final String DELIMITER_FOR_RESULT = " ";
+    private static final char LETTER = 'w';
+
     public String[] readFromFile(String fileName) {
-        //write your code here
-        return null;
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileName));
+            StringBuilder stringBuilder = new StringBuilder();
```
olekskov, line 16: please choose name that will indicate content of StringBuilder

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,42 @@
...
+            String value = reader.readLine();
+            if (value == null) {
+                return new String[] {};
+            }
+            while (value != null) {
+                stringBuilder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
+            String[] split = stringBuilder.toString().toLowerCase().split(DELIMITER);
```
olekskov, line 25: better to use nouns as names for variables
if this string were something like below it would be much more understandable:
```suggestion
            String[] fileContentByWords = fileContent.toString().toLowerCase().split(DELIMITER);
```

PR #893, opened 2023-07-13
jv-read-from-file-solution

literandltx:


literandltx:
Have a question, how correct declare a variable? In the last commit, you can see exactly which announcement it is about.
If do not change "mvn clean package" would fail build.

PR #1650, opened 2026-04-26
firt solution

Viodon:


src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,45 @@
...
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+
 public class FileWork {
     public String[] readFromFile(String fileName) {
-        //write your code here
-        return null;
+        File file = new File(fileName);
```
Elena-Bruyako, line 11: you don't need to create new File

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,45 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
Elena-Bruyako, line 14: yse try with resources with BufferedReader or other classes that implements AutoCloseable

src/main/java/core/basesyntax/FileWork.java
```diff
@@ -1,8 +1,45 @@
...
+                data[i] = data[i].replaceAll("[\\p{Punct}]+$", "");
+                counterW++;
+            }
+        }
+        if (counterW == 0) {
+            return new String[0];
+        }
+        String[] result = new String[counterW];
+        for (String world: data) {
+            if (world.startsWith("w")) {
```
Elena-Bruyako, line 37: create constant for symbols 
`private static final String STARTING_LETTER = "w";`

Elena-Bruyako (changes requested):
see comments

